PR #44, opened 2018-04-01
Fix parsing of database name from url

### LINKED ISSUE #33: Database name not correctly extracted from URL path.
Branch: nio

Swift Version: Swift 4.1 Snapshot 2018-03-16

When making a new Postgres database configuration using the database url in the format

`postgres://user:password@host:port/databaseName`

the database name extracted from the URL is prefixed with the "/" (which is expected behavior of of `URL(string: "some url")?.path)` and therefore the connection to the postgres database cannot be made successfully.

Tested same database config using `PostgreSQLDatabaseConfig(hostname: port: username: , database: password: )` and connection to database worked fine so I know the database name is correct.

mayur-mahajan:
Fixes #33 and adds some additional validation to url scheme as per https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNSTRING

### DIFF
--- a/Sources/PostgreSQL/Database/PostgreSQLDatabaseConfig.swift
+++ b/Sources/PostgreSQL/Database/PostgreSQLDatabaseConfig.swift
@@ -3,7 +3,7 @@ import Foundation
 public struct PostgreSQLDatabaseConfig {
     /// Creates a `PostgreSQLDatabaseConfig` with default settings.
     public static func `default`() -> PostgreSQLDatabaseConfig {
-        return .init(hostname: "localhost", port: 5432, username: "postgres")
+        return .init(hostname: "localhost", port: nil, username: "postgres")
     }
 
     /// Destination hostname.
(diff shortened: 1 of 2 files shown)


Sources/PostgreSQL/Database/PostgreSQLDatabaseConfig.swift
```diff
@@ -23,9 +23,9 @@ public struct PostgreSQLDatabaseConfig {
     public let password: String?
     
     /// Creates a new `PostgreSQLDatabaseConfig`.
-    public init(hostname: String, port: Int = 5432, username: String, database: String? = nil, password: String? = nil) {
+    public init(hostname: String, port: Int?, username: String, database: String? = nil, password: String? = nil) {
         self.hostname = hostname
-        self.port = port
+        self.port = port ?? 5432
```
twof: What's the reasoning behind this change? `port` being Optional doesn't make a whole lot of sense in my opinion because it doesn't accurately reflect the fact that a port is required.

mayur-mahajan: Since URLs allow for no port to be specified, I was motivated to keep parity across all initializers.
On second thought changing the signature will require changes elsewhere, outside the scope of what this PR is supposed to address. Will revert this bit.

twof: > Since URLs allow for no port to be specified
I didn't know that! Very interesting. 

Does that mean clients are expected to be able to connect to Postgres without a port specified in that case?

mayur-mahajan: https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNSTRING
Section `33.1.1.2. Connection URIs` seems to indicate so. Unfortunately there is no access to username from the environment otherwise `postgresql://` would have been valid as well.

PR #58, opened 2018-05-10
Fix Int32 -> UInt32 OIDs

mura-admin:
"The oid type is currently implemented as an unsigned four-byte integer" as per https://www.postgresql.org/docs/current/static/datatype-oid.html


penny-coin:
Hey @mura-admin, you just merged a pull request, have a coin! 

You now have 1 coins.

PR #68, opened 2018-05-22
Fix numeric with leading zero fractional

dmonagle:
Numeric numbers with a fractional part with a leading zero (eg: 10.08) will be decoded from the database without the leading zero (ie: 10.80).

This is a serious issue which results in corrupted numeric values.


penny-coin:
Hey @dmonagle, you just merged a pull request, have a coin! 

You now have 1 coins.

PR #75, opened 2018-06-15
An attempt to fix CircleCI configuration

Cellane:
Since the `gm` branch no longer exists, this should hopefully be the correct configuration… maybe?


Cellane:
Okay, I actually have no idea, sorry 😓

PR #95, opened 2018-07-26
Fix OID decoding as UInt32

ViliamKopecky:
OID doesn’t fit in so we had to bump it up to UInt32


penny-coin:
Hey @ViliamKopecky, you just merged a pull request, have a coin! 

You now have 1 coins.

PR #99, opened 2018-08-09
Allow custom binary operators

stjernegard:
Fixes https://github.com/vapor/postgresql/issues/98

There are a lot of binary operators in Postgres. Trying to catch them all in an enum seems very optimistic. Especially given that enums can't be extended with extra cases by the user.
Adding a `custom` case will allow the user to provide the operator in case the operator is not included in the enum already.

### DIFF
--- a/Sources/PostgreSQL/SQL/PostgreSQLBinaryOperator.swift
+++ b/Sources/PostgreSQL/SQL/PostgreSQLBinaryOperator.swift
@@ -1,195 +1,112 @@
 /// PostgreSQL specific `SQLBinaryOperator`.
-public enum PostgreSQLBinaryOperator: SQLBinaryOperator, Equatable {
+public struct PostgreSQLBinaryOperator: SQLBinaryOperator, Equatable, ExpressibleByStringLiteral {
     /// See `SQLBinaryOperator`.
-    public static var equal: PostgreSQLBinaryOperator { return ._equal }
-    
+    public static let add = PostgreSQLBinaryOperator("+")
+
     /// See `SQLBinaryOperator`.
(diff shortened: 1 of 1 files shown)


calebkleveter:
We should consider whether we want to add a `.custom` case to the enum, or just change the enum to a struct with static properties. We should be able to do that without any breaking changes, the syntax would be better, and I think it would be more future proof (not sure in what way, so maybe we can ignore that last point).

stjernegard:
I absolutely support using a struct instead. They're also much easier to extend and we don't risk breaking source compatibility when adding more cases in the future (on the off-chance that anyone has chosen to switch on this enum).

With the last part there, I guess this is _technically_ a breaking change.

I'll update my PR in a moment.

stjernegard:
If this change is accepted, it might be worth making the same change on vapor/sql.

https://github.com/vapor/sql/blob/master/Sources/SQL/SQLBinaryOperator.swift#L63

vzsg:
I'm also completely in favor of dropping the enum and replacing it with a struct with static, predefined cases.

MrLotU:
I'm in favour, but IMO it'd look cleaner to replace all inits with 
```swift
.init("data")
```
instead of
```swift
PostgreSQLBinaryOperator("data")
```

stjernegard:
@MrLotU, Unless I misunderstand you, I don't think that will compile.

`public static let add = .init("+")` does not compile, even when implementing an interface saying that there should be a `static let add: Self`.

MrLotU:
Ah, no, it wouldn't, I messed up. My bad 😅

penny-coin:
Hey @stjernegard, you just merged a pull request, have a coin! 

You now have 1 coins.

PR #111, opened 2018-11-02
Fix zeros stripped in conversion from numeric

michal-tomlein:
Fixes a bug causing zeros to be stripped from numeric values such as 10234.543201, producing 1234.54321.

### DIFF
--- a/Sources/PostgreSQL/Data/PostgreSQLData+String.swift
+++ b/Sources/PostgreSQL/Data/PostgreSQLData+String.swift
@@ -51,7 +51,7 @@ extension String: PostgreSQLDataConvertible {
                         integer += offset == 0 ? string : String(repeating: "0", count: 4 - string.count) + string
                     } else {
                         // Leading zeros matter with fractional
-                        fractional += fractional.count == 0 ? String(repeating: "0", count: 4 - string.count) + string : string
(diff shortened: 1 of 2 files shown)


Sources/PostgreSQL/Data/PostgreSQLData+String.swift
```diff
@@ -48,10 +48,10 @@ extension String: PostgreSQLDataConvertible {
                     
                     /// depending on our offset, append the string to before or after the decimal point
                     if offset < metadata.weight.bigEndian + 1 {
-                        integer += string
+                        integer += offset == 0 ? string : String(repeating: "0", count: 4 - string.count) + string
```
tanner0101: Breaking this up into one or two lines with some comments would make it a lot more readable I think

tanner0101:
I would love to include this change in 1.1.1. Let me know if you have any questions about writing the test. I'd also be happy to help write one.

michal-tomlein:
I can look into writing the test tomorrow (it’s late here). To be honest I don’t think a test should delay the merge of a critical data corruption bug fix. I could open a separate pull request.

penny-coin:
Hey @michal-tomlein, you just merged a pull request, have a coin! 

You now have 1 coins.

PR #117, opened 2018-12-19
Fix conversion between Date and timestamp/timestamptz types

michal-tomlein:
It seems `timestamp` values are adjusted for the local time zone.
When converting to a `Date`, we need to subtract the local offset to get the correct value.
When converting from a `Date`, it is better to use the `timestamptz` type, which works with both kinds of fields and does not have this problem.

### DIFF
--- a/Sources/PostgreSQL/Data/PostgreSQLData+Date.swift
+++ b/Sources/PostgreSQL/Data/PostgreSQLData+Date.swift
@@ -11,7 +11,12 @@ extension Date: PostgreSQLDataConvertible {
             }
         case .binary(let value):
             switch data.type {
-            case .timestamp, .timestamptz:
+            case .timestamp:
+                let microseconds = value.as(Int64.self, default: 0).bigEndian
+                var seconds = Double(microseconds) / Double(_microsecondsPerSecond)
+                seconds -= Double(TimeZone.current.secondsFromGMT())
+                return Date(timeInterval: seconds, since: _psqlDateStart)
+            case .timestamptz:
                 let microseconds = value.as(Int64.self, default: 0).bigEndian
                 let seconds = Double(microseconds) / Double(_microsecondsPerSecond)
                 return Date(timeInterval: seconds, since: _psqlDateStart)
@@ -28,7 +33,7 @@ extension Date: PostgreSQLDataConvertible {
 
     /// See `PostgreSQLDataConvertible`.
     public func convertToPostgreSQLData() throws -> PostgreSQLData {
-        return PostgreSQLData(.timestamp, binary: Data.of(Int64(self.timeIntervalSince(_psqlDateStart) * Double(_microsecondsPerSecond)).bigEndian))
+        return PostgreSQLData(.timestamptz, binary: Data.of(Int64(self.timeIntervalSince(_psqlDateStart) * Double(_microsecondsPerSecond)).bigEndian))
     }
 }
 
--- a/Tests/PostgreSQLTests/PostgreSQLConnectionTests.swift
+++ b/Tests/PostgreSQLTests/PostgreSQLConnectionTests.swift
@@ -607,6 +607,47 @@ class PostgreSQLConnectionTests: XCTestCase {
         }
     }
 
+    // Stores the current date in timestamp and timestamptz fields and compares it with the result of now()
+    func testTimestampDecode() throws {
+        enum Error: Swift.Error {
+            case noRowsFound
+        }
+
+        let conn = try PostgreSQLConnection.makeTest()
+        struct TimestampTest: PostgreSQLTable, Equatable {
+            static let sqlTableIdentifierString = "timestamptest"
+            var ts: Date // timestamp
+            var tstz: Date // timestamptz
+            var tsNow: Date? // now() as timestamp
+            var tstzNow: Date? // now() as timestamptz
+        }
+
+        defer { try? conn.drop(table: TimestampTest.self).ifExists().run().wait() }
+        try conn.create(table: TimestampTest.self)
+            .column(for: \TimestampTest.ts, type: .timestamp)
+            .column(for: \TimestampTest.tstz, type: .timestamptz)
+            .column(for: \TimestampTest.tsNow, type: .timestamp, .default(.function(.function("now", []))))
+            .column(for: \TimestampTest.tstzNow, type: .timestamptz, .default(.function(.function("now", []))))
+            .run().wait()
+
+        let date = Date()
+
+        let test = TimestampTest(ts: date, tstz: date, tsNow: nil, tstzNow: nil)
+        try conn.insert(into: TimestampTest.self).value(test).run().wait()
+        let fetched = try conn.select().all().from(TimestampTest.self).first(decoding: TimestampTest.self).unwrap(or: Error.noRowsFound).wait()
+
+        let timeInterval = date.timeIntervalSinceReferenceDate
+        let ts = fetched.ts.timeIntervalSinceReferenceDate
+        let tstz = fetched.tstz.timeIntervalSinceReferenceDate
+        let tsNow = fetched.tsNow?.timeIntervalSinceReferenceDate ?? 0.0
+        let tstzNow = fetched.tstzNow?.timeIntervalSinceReferenceDate ?? 0.0
+
+        XCTAssert(abs(ts - timeInterval) < 1.0)
+        XCTAssert(abs(tstz - timeInterval) < 1.0)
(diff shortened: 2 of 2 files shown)


michal-tomlein:
Here are the results I’m getting in `testTimestampDecode` without this patch:

For a `date` value of `566990074.20886695` (`2018-12-20 09:14:34 UTC`), I got:
`ts == 566990074.20886695` (`2018-12-20 09:14:34 UTC`)
`tstz == 566986474.20886695` (`2018-12-20 08:14:34 UTC`)
`tsNow == 566993674.21106505` (`2018-12-20 10:14:34 UTC`)
`tstzNow == 566990074.21106505` (`2018-12-20 09:14:34 UTC`)

michal-tomlein:
In my case, this bug manifests itself in the following way: I have a `Token` model with an `expirationDate: Date` property (`timestamptz` in the database), which is set to `Date(timeIntervalSinceNow: 300.0)` on init. In GMT+1, this actually creates expired tokens.

Because `timestamptz` and `Date` are both representations of a point in time, it makes sense to use `.timestamptz` when converting to `PostgreSQLData`. This fixes the bug.

I’m a little less clear on what to do about `timestamp`, which is a representation of a wall clock. This is not a good match for storing `Date` values. My patch makes it work so that you get back exactly what you put in. I think this is the least bad option.

MrMage:
We are affected by the same bug. Thank you for this fix! I do think that the fix only works as long as the Vapor client's and the Postgres server's timezones match, however.

@tanner0101, could you please take a look at this?

penny-coin:
Hey @michal-tomlein, you just merged a pull request, have a coin! 

You now have 2 coins.

MrMage:
@tanner0101 thanks for merging; however, I think this will still lead to issues when the Vapor binary and the Postgres server are not set to the same timezones. /cc @michal-tomlein

michal-tomlein:
@MrMage, that should only be the case if you’re using `timestamp`, which is conceptually incompatible with `Date`. `timestamptz` should always work correctly.

MrMage:
@michal-tomlein but your code explicitly references the current Foundation timezone and subtracts that. If the Postgres server expects a different offset to be subtracted (because it is in a different timezone), wouldn't this break down?

MrMage:
(Looking at this more, im probably misunderstanding something, sorry.)

michal-tomlein:
@MrMage, yes, it would break down, but that code is only in `case .timestamp`, not `case .timestamptz` (below), where such offset subtraction is not necessary.

MrMage:
Sorry, thinking about this more I know understand that this should work correctly with `timestamptz`. However, I have three questions, maybe @tanner0101 could provide some insight:

1. Given that this will the behavior of existing code, should this be considered a breaking change, even though it only "fixes" the  previously-broken behavior?
2. Will auto migrations currently create `timestamp` or `timestamptz` for `Date` columns?
3. Given that `timestamp` can cause really weird behavior when interpreted as a date and yields different results when used in different timezones, should that type maybe decoded as `Int64` rather than `Date`?

michal-tomlein:
Good points. I’d argue auto migrations should be changed to create `timestamptz` fields if they don’t already.

Re 3., I think `String` would work better. `Int64` has all the same problems as `Date`, which can be thought of as just a wrapper around a `Double`.

tanner0101:
1: No, I don't think so. Bug fixes should be patch releases. 
2: Looks like [it's using `.timestamp`](https://github.com/vapor/postgresql/blob/1/Sources/PostgreSQL/SQL/PostgreSQLDataTypeStaticRepresentable.swift#L21). I agree that it _should_ be using `.timestamptz`. 
3: The model definition controls which type the data is decoded as. If you store `timestamp: Int` on your model, then an `Int` will be decoded.

mikkelu:
Updating to this has broken all my Vapor applications that rely on `timestamp` and foundation `Date()` implementations. 

After looking more into it, I agree that it should be `.timestamptz`, however, this change would require me to migrate my database, and therefore think this PR is breaking change.

tanner0101:
@mikkelu can you elaborate on what you mean by broken?

mikkelu:
I have an access token that stores an expiry field (in the timestamp format). When checking if this is valid, it will error out with an expired error. 

I have something like this:
```
var isValid: Bool {
  print(Date())
  print(expiry)
  return expiry >= Date()
}
```

Here is the result with my database as-is: 
`Date() => 2019-03-01 19:57:50 +0000`
`expiry => 2019-03-01 19:08:57 +0000`

And here is the result if a do a quick manual migration to `timestamptz`:
`Date() => 2019-03-01 19:58:44 +0000`
`expiry => 2019-03-01 20:08:57 +0000`

As you can see, it will require me to migrate my database. Unless there is something I am missing?

mikkelu:
I have an access token that stores an expiry field (in the timestamp format). When checking if this is valid, it will error out with an expired error. 

I have something like this:
```
var isValid: Bool {
  print(Date())
  print(expiry)
  return expiry >= Date()
}
```

Here is the result with my database as-is: 
`Date() => 2019-03-01 19:57:50 +0000`
`expiry => 2019-03-01 19:08:57 +0000`

And here is the result if a do a quick manual migration to `timestamptz`:
`Date() => 2019-03-01 19:58:44 +0000`
`expiry => 2019-03-01 20:08:57 +0000`

As you can see, it will require me to migrate my database. Unless there is something I am missing?

tanner0101:
Ah, I see now. Yeah, @MrMage was right this probably should be considered breaking. I'll delete the tag to remedy this now, then we can discuss our best option for fixing.

tanner0101:
The tag has been deleted. @mikkelu do a clean build (revert package.resolved) and you should be back to normal.

tanner0101:
Ideas for how to merge this:
- Minor tag w/ fix. 
    - Tag the fix as `1.5.0`
    - Include a temporary helper struct like `_PostgreSQLDate` for re-creating the old behavior models. 
    - Include a SQL query for converting DBs to use `TIMESTAMPTZ`
- Include in 2.0
    - Push this fix to the next major release.

Either way, we should probably change `Date`'s default type to `TIMESTAMPTZ` sooner than later to avoid more people needing to migrate their DBs in the future.

mikkelu:
Thanks, @tanner0101 
I agree that `TIMESTAMPTZ` should be used as soon as possible. It makes more sense. 
I think it would be a great idea to include that SQL query that converts timestamps, so people can quickly adopt this.

MrMage:
FYI, migrating Date to timestamptz should not be done without this fix, either, because currently writing a date to the DB, then reading it back from the DB shifts it by the databases timezone offset. I can provide sample code to illustrate this behavior on Monday.

So both changes (fixing this and switching Date to timestamptz) should probably be released at the same time.

tanner0101:
Ah yeah, @MrMage is right. I think need the change to happen in all 3 places at once:
- Switch `Date`'s preferred column to TIMESTAMPTZ
- Switch `Date` to encoding `.timestamptz`
- Fix `Date`'s decoding of `timestamp`

Although, _maybe_ we could get away with:
- Switch `Date`'s preferred column to `TIMESTAMPTZ`
- Switch `Date` to encoding `.timestamptz`
- Leave `Date`'s decoding of `timestamp` broken. 

Unless I'm missing something, the second option could at least prevent new code from succumbing to the problem without breaking old code.

MrMage:
I'd love if the second option worked, but I think it won't: with the second option old data would be decoded "correctly" because decoding is broken in the same way. However, new data would be encoded correctly, leading to a broken decoding (because the decoding is still broken), plus we would be inconsistent across old and new data.

tanner0101:
Yeah, I fear you're right. We'll probably need to err on the safe side and change everything at once. Now the question becomes: `1.5.0` or `2.0.0`?

MrMage:
I tend towards 2.0 and possibly even logging a runtime warning once per launch from 1.5.0 onwards if badly-encoded Dates are being used. This could cause really nasty surprises when migrating to 2.0 without being aware of it.

MrMage:
For posterity, this is an exemplary test I used to determine that Vapor's encoding/decoding prior to this patch would cause errors when writing to a `timestamptz` column:

```
	func testCreateUser() {
		let timestamp = Date()
		let user = User(email: "email", passwordHash: "password", signupDate: timestamp)
		_ = try! user.create(on: connection).wait()
		
		let user2 = try! User.fetchByID(2, context: context).wait()!
		XCTAssertEqual(timestamp, user2.signupDate)  // fails for Postgres server timezones other than UTC
	}
```

PR #90, opened 2018-07-12
Fix regexp operator

ffried:
This fixes the RegExp operator. PostgreSQL uses `~` and not `REGEXP`.

PR #139, opened 2019-08-06
Fix not-regexp operator, add case insensitive regexp operators

ffried:
Fixes the `notRegexp` operator and adds the two case-insensitive variants (#139, #90).

See [the pgsql docs](https://www.postgresql.org/docs/10/functions-matching.html#FUNCTIONS-POSIX-REGEXP) for reference.


ffried:
@tanner0101 Ping... 🙂

ffried:
No worries. Thanks for merging!

PR #142, opened 2019-09-08
Handle NULLs when decoding arrays

### LINKED ISSUE #141: Fatal error when decoding an array with a NULL value in it
Apparently, Postgres encodes NULL values inside arrays in a way that PostgreSQLDataDecoder does not anticipate. In this case, the byte `count` of the value is -1, which causes [this line](https://github.com/vapor/postgres-kit/blob/1/Sources/PostgreSQL/Utilities/Utilities.swift#L167) called from [here](https://github.com/vapor/postgres-kit/blob/1/Sources/PostgreSQL/Codable/PostgreSQLDataDecoder.swift#L66) to explode with a friendly error:

> Fatal error: Can't form Range with upperBound < lowerBound

---

While this originally was discovered by using the `array_agg` function combined with left joins (which triggers [this weird behavior](https://stackoverflow.com/questions/31108946/postgres-returns-null-instead-of-for-array-agg-of-join-table)), there is a much simpler reproduction:

```swift
conn.raw("SELECT ARRAY[NULL, 'foo', NULL, 'bar'] AS arr")
    .all(decoding: SimpleArray.self)
```

where SimpleArray is defined as:

```swift
struct SimpleArray: Decodable {
    let arr: [String?]
}
```

---

The issue was discovered with Vapor 3 / Postgres (driver) 1.4.1.

V4 doesn't seem to be affected as array decoding is [not implemented at all](https://github.com/vapor/postgres-kit/blob/74d49fd4dcd9e2e47bb56f63ebaf27ea60706a6b/Sources/PostgresKit/PostgresDataDecoder.swift#L38) at the time of writing.

vzsg:
Tests and fixes #141.

### DIFF
--- a/Sources/PostgreSQL/Codable/PostgreSQLDataDecoder.swift
+++ b/Sources/PostgreSQL/Codable/PostgreSQLDataDecoder.swift
@@ -63,6 +63,13 @@ struct PostgreSQLDataDecoder {
                     let _ = value.extract(Int32.self).bigEndian
                     for _ in 0..<count {
                         let count = Int(value.extract(Int32.self).bigEndian)
+
+                        // Postgres sends -1 to distinguish an empty byte sequence from NULLs
+                        guard count != -1 else {
+                            array.append(.null)
+                            continue
+                        }
+
                         let subValue = value.extract(count: count)
                         let psqlData = PostgreSQLData(type, binary: subValue)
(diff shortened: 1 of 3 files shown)


penny-coin:
Hey @vzsg, you just merged a pull request, have a coin! 

You now have 2295 coins.

PR #203, opened 2021-01-15
Fix passing the logger through on `PostgresConnectionSource`

joshuawright11:
Pass the correct logger through to `PostgresConnectionSource` to bring it in line with [MySQLKit](https://github.com/vapor/mysql-kit/blob/master/Sources/MySQLKit/MySQLConnectionSource.swift#L21).


0xTim (approved):
LGTM - I'll let @gwynne have the final work in case this was done for a reason

joshuawright11:
Just following up. Any thoughts @gwynne or @0xTim ?

0xTim:
@joshuawright11 could you push an empty commit to retrigger CI?

joshuawright11:
@0xTim done. Looks like you need to approve running the workflow.

PR #211, opened 2021-09-23
Address issues in decoding data from the database

gwynne:
Basically this just cleans up `PostgresDataDecoder` quite a bit, along with the `PostgresKit` unit tests, such as they are. It's difficult to say whether this will have any performance impact (probably not).


0xTim (approved):

gwynne:
Okay, all the test failures now are just the stupid performance metric being stubborn about outlier values caused by the SCRAM-SHA-256 auth. Does anyone care? Can I just merge this? 🤣

Joannis (approved):

Issue #207, opened 2021-05-27
Crash in `_KeyedEncoder` when encoding a "SE-0295" `Codable`


finestructure:
### Describe the bug

I've defined a `Codable` enum with associated types with custom encoding to match Swift 5.5's format of `Codable` synthesis (SE-0295). The idea is that I can start storing data with 5.4 and then remove the custom `Codable` conformance when 5.5 ships.

When trying to store a value of this type to a Postgres database, the `save` crashes in `_KeyedEncoder`, trying to encode the data.

NB: the fact that I'm trying to match SE-0295's format may be a red herring, I've not tried a different format!

### To Reproduce

I'm attaching a standalone Swift playground that reproduces the crash and shows that encoding with `JSONEncoder` works as expected.

### Expected behavior

Encoding of a value that's encodable with `JSONEncoder` should not crash.

### Environment

Xcode 12.5

[postgres-kit Playground.zip](https://github.com/vapor/postgres-kit/files/6551794/postgres-kit.Playground.zip)

siemensikkema:
@finestructure Thanks for making this issue so easy to reproduce! It appears that `PostgresDataEncoder` does not support summoning nested containers (it just `fatalError`s). I don't have the full overview of whether it should but I can provide a workaround. By conforming `ProductType` to `PostgresJSONCodable` the whole encoding step is short-circuited and you should get the expected result.

finestructure:
Hi @siemensikkema , thanks a lot for taking a look!

There's probably also another way to work around this issue. I haven't tried it in this instance yet but I'm pretty sure we have older encodable enums with associated types where we used

```
"key": null
```

to encode presence of keys and keys with values rather than nested containers. The reason I went for the new format is that with [SE-0295 in Swift 5.5](https://github.com/apple/swift-evolution/blob/main/proposals/0295-codable-synthesis-for-enums-with-associated-values.md) I believe the standard auto-generated format is going to be [the nested one](https://github.com/apple/swift-evolution/blob/main/proposals/0295-codable-synthesis-for-enums-with-associated-values.md#structure-of-encoded-enums).

I.e. any user trying to store an enum with associated types using `Codable` synthesis will probably hit this exception.

I'll try your suggested workaround as that should allow me to retain the format and then I can just drop the `PostgresJSONCodable` should support be added. I'll report back how it goes!

finestructure:
Just to follow up: making `ProductType` conform to `PostgresJSONCodable` allows saving but not reading apparently:

[...]

PR #216, opened 2022-01-20
Add support for nested JSON

tetraoxygen:
This PR adds support for encoding nested JSON. For example, you can encode a Codable datatype that includes other Codable types.

You can use this with Codable's polymorphism support to add small bits of data that support polymorphism by having a type annotation outside a nested object you intend to decode.

### DIFF
--- a/Sources/PostgresKit/PostgresDataDecoder.swift
+++ b/Sources/PostgresKit/PostgresDataDecoder.swift
@@ -40,14 +40,11 @@ public final class PostgresDataDecoder {
 
     enum Error: Swift.Error, CustomStringConvertible {
         case unexpectedDataType(PostgresDataType, expected: String)
-        case nestingNotSupported
 
         var description: String {
             switch self {
             case .unexpectedDataType(let type, let expected):
                 return "Unexpected data type: \(type). Expected \(expected)."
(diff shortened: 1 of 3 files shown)


0xTim (commented):
@fabianfett any thoughts?

@tetraoxygen thanks for this! We'll need tests for the new behaviour at the very least

tetraoxygen:
I've added tests for nested encoding, which is all I materially changed. I'm also happy to go write nested decode tests (though those just use Swift's standard `JSONDecoder`) if it would be useful.

tetraoxygen:
@0xTim Following up on this — should I write more tests or do the ones I've added do the job?

tetraoxygen:
I believe #207 would be solved by this PR (without the PostgresJSONCodable workaround)

0xTim (changes requested):
Could you set the spacing to 4 spaces instead of a tab to keep it consistent with the rest of the codebase?

@gwynne @fabianfett any comments?

tetraoxygen:
Indentation should be fixed now

sidepelican:
I have also this trouble. Thanks for the nice PR.

gwynne:
This should be fixed with the new logic in #241

PR #218, opened 2022-02-04
Ignore ParamaterStatus messages when handling response

baarde:
ParameterStatus messages may be generated in response to a SET command or when the administrator sends the SIGHUP signals to the server. They will appear in the query response and should be ignored.


baarde:
### Underlying issue

When sending the following command:
```sql
SET TIME ZONE 'UTC';
```

or after sending `SIGHUP` to the server, we get the following error:
```
[PostgreSQLError.query: Unexpected message during `PostgreSQLParseRequest`: parameterStatus(TimeZone: UTC)]
```

This issue doesn't affect PostgresKit version 2.

0xTim:
This was released in [1.7.0](https://github.com/vapor/postgres-kit/releases/tag/1.7.0)

PR #225, opened 2022-06-03
Allow `backendKeyData` to be set

rausnitz:
This change provides access to the new functionality provided in this postgres-nio PR: https://github.com/vapor/postgres-nio/pull/296. In short, apps can enable integration with Amazon RDS Proxy by setting `requireBackendKeyData` to false. See the postgres-nio PR for more details.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -10,7 +10,7 @@ let package = Package(
         .library(name: "PostgresKit", targets: ["PostgresKit"]),
     ],
     dependencies: [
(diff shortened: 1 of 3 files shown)


rausnitz:
Checks using Swift 5.2 are failing because dependencies have dropped support for Swift 5.2. Do we need to modify the `swift-tools-version` in this library, modify the checks, or both?

0xTim:
@rausnitz we'll fix this in #226 and then you can pull that in once merged

rausnitz:
Should be good to go now

0xTim (approved):

PR #231, opened 2022-08-18
Migrate to Swift Atomics

fatto:
Fix deprecation warnings from `NIOAtomic` and migrate any uses to `swift-atomics`


0xTim (approved):
Thanks!

0xTim:
Ignoring the API breaking change failure because it's wrong

PR #233, opened 2022-10-19
Update CI to include Swift 5.7 and Postgres 15

gwynne:
Haven't yet built a working reusable workflow that would allow centralizing this for packages which require supporting services (such as a database server).


0xTim (approved):
LGTM - I'm assuming there are no major changes between 11-15 that would necessitate us testing individual features (and even if there are we should probably push that into PostgresNIO)

gwynne:
Nope, at least nothing that PostgresKit cares about - that's all PostgresNIO's concern, and even then the only real wire protocol change since 11 that I can think of offhand was the addition of streaming result rows.

Issue #232, opened 2022-10-14
Connection issue while URL contains "%"


shawnclovie:
### Describe the bug

While my connection URL contains "%", e.g. `postgres://user:2kf%D@127.0.0.1:5432/db1` -
* `URL(string:)` would return nil
`PostgresConfiguration(url:)` would failed, `DatabaseConfigurationFactory.postgres()` too.

* If I change the URL, make % as %25
connection could be ok, but password is incorrect since `PostgresConfiguration(url:)` taks url.password and was not call `removingPercentEncoding`, actual password sent is `2kf%25D` for this example.

* My workaround
```swift
func makePostgresConnection(url: URL) {
	guard let host = url.host,
		let user = url.user?.removingPercentEncoding,
		let pass = url.password?.removingPercentEncoding else {
		throw Error("informal connect string, no host or user or password")
	}
	let pgcfg = PostgresConfiguration(
		hostname: host,
		port: url.port ?? PostgresConfiguration.ianaPortNumber,
		username: user,
		password: pass,
		// two lines below were copied from PostgresConfiguration(url:)
		database: url.path.split(separator: "/").last.flatMap(String.init),
		tlsConfiguration: url.query?.contains("ssl=true") == true || url.query?.contains("sslmode=require") == true ? .makeClientConfiguration() : nil)
	return DatabaseConfigurationFactory.postgres(configuration: pgcfg, ...).make()
}
```

### Expected behavior
* Use `removingPercentEncoding` while take user and password from the URL.
* Find batter way convert url String to URL for password contains "%".

elGeekalpha:
#235

gwynne:
This was solved by #241

PR #235, opened 2022-11-07
Add postgress configuration encoding support

elGeekalpha:
PR solves #232 validating url with from PostgresConfiguration with urlQueryAllowed CharacterSet to allow passing passwords with special characters like `postgres://user:se%cret@localhost`

Also aded some test to verify PostgresConfiguration init with several urls configurations and init components validations.


gwynne:
This has been addressed by proper URL handling in the new `SQLPostgresConfiguration` implementation.

PR #245, opened 2023-06-09
Remove the temporary workaround for PSQLError descriptions

gwynne:
The workaround didn't work very well anyway (I got it slightly wrong), and thanks to vapor/postgres-nio#372 is no longer needed at all.

This reverts #244, commit 875e8c191ea776802fddf6773146349803629754.


tkrajacic (approved):
Looks good to me.

PR #249, opened 2023-08-11
Remove use of legacy codepath when establishing Postgres connections

gwynne:
Also tidies CI as usual.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,100 +27,120 @@ jobs:
   api-breakage:
     if: ${{ !(github.event.pull_request.draft || false) }}
     runs-on: ubuntu-latest
-    container: swift:5.8-jammy
+    container: swift:jammy
     steps:
       - name: Check out code
         uses: actions/checkout@v3
         with: { 'fetch-depth': 0 }
       - name: Run API breakage check action
-        uses: vapor/ci/.github/actions/ci-swift-check-api-breakage@reusable-workflows
+        uses: vapor/ci/.github/actions/ci-swift-check-api-breakage@main
 
-  # Run Linux unit tests against various configurations
+  code-coverage:
+    if: ${{ !(github.event.pull_request.draft || false) }}
+    runs-on: ubuntu-latest
+    container: swift:jammy
+    services:
+      psql-a:
+        image: postgres:15
+        env:
+          POSTGRES_USER: test_username
+          POSTGRES_DB: test_database
+          POSTGRES_PASSWORD: test_password
+          POSTGRES_HOST_AUTH_METHOD: scram-sha-256
+          POSTGRES_INITDB_ARGS: --auth-host=scram-sha-256
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+      - name: Run unit tests for coverage data
+        run: swift test --enable-code-coverage
+      - name: Upload coverage data
+        uses: vapor/swift-codecov-action@v0.2
+  
+  gh-codeql:
+    if: ${{ !(github.event.pull_request.draft || false) }}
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -27,100 +27,120 @@ jobs:
   api-breakage:
     if: ${{ !(github.event.pull_request.draft || false) }}
     runs-on: ubuntu-latest
-    container: swift:5.8-jammy
+    container: swift:jammy
```
MahdiBM, line 30: Why this change? It's kind of inconsistent with all other CIs.

.github/workflows/test.yml
```diff
@@ -133,43 +153,31 @@ jobs:
...
+          xcode-version: ${{ matrix.xcode-version }}
       - name: Install Postgres, setup DB and auth, and wait for server start
-        env:
-          POSTGRES_VERSION: ${{ matrix.dbimage }}
-          POSTGRES_AUTH_METHOD: ${{ matrix.dbauth }}
         run: |
-          export PATH="$(brew --prefix)/opt/${POSTGRES_VERSION}/bin:$PATH" PGDATA=/tmp/vapor-postgres-test
-          (brew unlink postgresql || true) && brew install "${POSTGRES_VERSION}" && brew link --force "${POSTGRES_VERSION}"
-          initdb --locale=C --auth-host "${POSTGRES_AUTH_METHOD}" -U "${POSTGRES_USER}" --pwfile=<(echo "${POSTGRES_PASSWORD}")
+          export PATH="$(brew --prefix)/opt/postgresql@14/bin:$PATH" PGDATA=/tmp/vapor-postgres-test
```
MahdiBM, line 175: why `@14`? Isn't the CI using postgres 15 now?

.github/workflows/test.yml
```diff
@@ -27,100 +27,120 @@ jobs:
...
+    container: swift:jammy
     steps:
       - name: Check out code
         uses: actions/checkout@v3
         with: { 'fetch-depth': 0 }
       - name: Run API breakage check action
-        uses: vapor/ci/.github/actions/ci-swift-check-api-breakage@reusable-workflows
+        uses: vapor/ci/.github/actions/ci-swift-check-api-breakage@main
+
+  code-coverage:
```
MahdiBM, line 38: This one is just for my curiosity as you probably already considered it, but why not use the reusable CI to get rid of separate `code-coverage` and `gh-codeql` jobs?

MahdiBM (dismissed):
The `@14`s seem suspicious 🤔

.github/workflows/test.yml
gwynne: It's in line with https://github.com/vapor/ci/blob/main/.github/workflows/run-unit-tests.yml#L56 and https://github.com/vapor/ci/blob/main/.github/workflows/run-unit-tests.yml#L70 - these jobs don't care overmuch about which Swift version they use in particular, it should always just be the "latest stable" version, which is exactly what the non-versioned Docker tags provide.

.github/workflows/test.yml
gwynne: The Linux unit tests cover Postgres versions 11, 13, and 15; we take the macOS tests as an excuse to cover 14 as well.

.github/workflows/test.yml
gwynne: Same reason as the other non-SQLite Fluent repos don't - the reusable workflow doesn't yet support the services specifications these repos need and I haven't yet made the reusable workflow capable of invoking "only" the CodeQL/coverage/API breakage jobs without any tests.

MahdiBM (approved):

PR #252, opened 2023-08-29
Temporarily pin PostgresNIO to 1.17.0

gwynne:
Per vapor/postgres-nio#405 et al.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -13,7 +13,7 @@ let package = Package(
         .library(name: "PostgresKit", targets: ["PostgresKit"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/vapor/postgres-nio.git", from: "1.14.2"),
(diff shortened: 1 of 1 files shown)


0xTim (approved):

Package.swift
```diff
@@ -13,7 +13,7 @@ let package = Package(
         .library(name: "PostgresKit", targets: ["PostgresKit"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/vapor/postgres-nio.git", from: "1.14.2"),
+        .package(url: "https://github.com/vapor/postgres-nio.git", exact: "1.17.0"),
```
MahdiBM, line 16: I would've used an upper-bound of 1.17.0 instead of "exact".
Someone might want to have their PostgresNIO pinned to a lower version (believe me, i've seen those, mostly as a mistake, but anyway)

0xTim: The other issue this might cause is people pinning to old versions for the error stuff as well

PR #253, opened 2023-10-29
[WIP] Performant decoding/encoding with updated PostgresNIO APIs

MahdiBM:
**This PR has been opened so we can discuss the direction of the change while i'm working on it. Nothing is final.**

Adds a macro name `PostgresRecord` for decoding/encoding of Swift types with newer more-performant PostgresNIO APIs.

### Example

```swift
@PostgresRecord
struct MyTable {
    let int: Int
    let string: String
}

func decode(from row: PostgresRow) -> MyTable {
    row.decode() /// The macro will handle the decoding from here.
}
```
The macro-generated code:
```swift
extension MyTable: PostgresRecord {
    init(
        _from row: PostgresRow,
        context: PostgresDecodingContext<some PostgresJSONDecoder>,
        file: String,
        line: Int
    ) throws {
        let decoded = try row.decode(
            (Int, String).self,
            context: context,
            file: file,
            line: line
        )
        self.int = decoded.0
        self.string = decoded.1
    }
}
```

### Checklist
- [ ] Handle encoding too.
  - [ ] Support snake-case key strategy.
- [ ] Properly document APIs as soon as we're settled on the APIs. 
- [ ] Tests for usage of the new APIs.
- [ ] Conditional Swift 5.9 support if we don't want to move to Swift 5.9+ just yet.


gwynne:
I'm confused, did you for some reason miss the fact that I updated PostgresKit to the newer APIs a _long_ time ago?

MahdiBM:
Nope, not really. This is just an inherent limitation of the newer PostgresNIO APIs, if you want them to be as performant as they should be (e.g. not use PostgresRandomAccessRow). You can't decode any arbitrary row with the newer APIs to an object/struct. You need to manually write out the tuple, and then one by one initialize your struct members.
This PR's primary purpose is to solve that.

PR #257, opened 2024-01-28
Fix mishandling of URL

### LINKED ISSUE #255: URL is not supported as a postgres codable type
### Describe the bug

URL doesn't automatically work as a field type with underlying database type 'string'. It can't encode the URL properly.

### To Reproduce

Have a model with this field:

```
@Field(key: "link")
var link: URL
```

And a migration like this:
```
func prepare(on database: Database) async throws {
      try await database.schema("bookmarks")
          .id()
          .field("link", .string, .required)
          .create()
  }
```

Then when you try to create one of these and save it to the database, you get this error:
```
invalid field: 'link', type: URL, error: Swift.DecodingError.dataCorrupted(Swift.DecodingError.Context(codingPath: [SomeCodingKey(stringValue: "bookmarks_link (2)", intValue: nil)], debugDescription: "Unable to interpret value of PSQL type TEXT: ByteBuffer { readerIndex: 0, writerIndex: 23, readableBytes: 23, capacity: 23, storageCapacity: 2048, slice: _ByteBufferSlice { 197..<220 }, storage: 0x000000011f00c200 (2048 bytes) }", underlyingError: nil)) [request-id: E5B62897-697A-468F-94A7-F6B5726CC7FB]
```

### Expected behavior

The model and field are saved properly to the database.

### Environment

* Vapor Framework version: 4.86.2
* Vapor Toolbox version: 18.7.4
* OS version: macOS 14.1 (23B74)
* Fluent version: 4.8.0
* fluent-postgres-driver version: 2.8.0

gwynne:
We have for some time now been incorrectly encoding `URL`s in an incorrect format that fails to trivially round-trip back through decoding. This is now fixed. Any previously broken data stored in a database affected by this issue will now be correctly read back and will have the correct format when next written.

> [!NOTE]
> This is **_not_** related in any way to the recent issues with Vapor's `URI` type; it is a separate concern affecting the `URL` type that comes from Foundation.

Also incidentally fixes #255


.github/dependabot.yml
```diff
@@ -0,0 +1,22 @@
+version: 2
+enable-beta-ecosystems: true
```
MahdiBM, line 2:
```suggestion

```

.github/dependabot.yml
```diff
@@ -0,0 +1,22 @@
...
+    directory: "/"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 6
+    allow:
+      - dependency-type: all
+    groups:
+      all-dependencies:
+        patterns:
+          - "*"
```
MahdiBM, line 22: We don't commit `Package.resolved` in this repo since it's just a library, so it there a point?

```suggestion
```

Package@swift-5.9.swift
```diff
@@ -0,0 +1,49 @@
+// swift-tools-version:5.9
+import PackageDescription
+
+let swiftSettings: [SwiftSetting] = [
+    .enableUpcomingFeature("ExistentialAny"),
+    .enableUpcomingFeature("ConciseMagicFile"),
+    .enableUpcomingFeature("ForwardTrailingClosures"),
+    .enableUpcomingFeature("DisableOutwardActorInference"),
+    .enableExperimentalFeature("StrictConcurrency=complete"),
+]
```
MahdiBM, line 10: It recently came to my attention that you can define this variable at the end of the file, but it needs to be computed instead of stored.

I'm not sure if you want to keep `ExistentialAny` since it's going to be removed. As i've mentioned before i do like it.

I'd also use `StrictConcurrency` instead of `StrictConcurrency=complete` since Holly removed the option to choose `targeted` from future Swift versions (she added a new `upcomingFeature` flag for it, instead of `experimental`, but it'snot available in 5.9)

Tests/PostgresKitTests/PostgresKitTests.swift
```diff
@@ -224,20 +224,30 @@ final class PostgresKitTests: XCTestCase {
         XCTAssertNoThrow(numericValue = try PostgresDataTranslation.decode(Double.self, from: .init(bytes: numericBuffer, dataType: .numeric, format: .binary, columnName: "", columnIndex: -1), in: .default))
         XCTAssertEqual(numericValue, Double(Decimal(12345.6789).description))
     }
+    
+    func testURLWorkaroundDecoding() throws {
```
MahdiBM, line 228: +1

MahdiBM (changes requested):

.github/dependabot.yml
gwynne: Yes.

Package@swift-5.9.swift
gwynne: I prefer `ExistentialAny` syntax. It's not going to be removed, it's just not going to become "required" until Swift 7 (according to [Features.def](https://github.com/apple/swift/blob/main/include/swift/Basic/Features.def#L130).

As for `StrictConcurrency`, it does no harm to specify `complete` for now. When I can switch to the `upcoming` version, I'll stop bothering.

.github/dependabot.yml
MahdiBM: What then? 😶‍🌫️

gwynne: Ya know what? Anything to make you happy, you bloody annoying person who's learned far too well from me and would be impressing me with the nitpickiness if I weren't being deliberately over the top ornery 😉

MahdiBM (approved):

PR #259, opened 2024-03-24
Apply updates for improved interop with SQLKit

gwynne:
These changes bump the minimum Swift version to 5.8 and prepare PostgresKit for [SQLKit's overhaul PR](https://github.com/vapor/sql-kit/pull/172). The SQLKit changes are _NOT_ required by this update.


README.md
```diff
@@ -24,8 +24,7 @@ The table below shows a list of PostgresKit major releases alongside their compa
 
 |Version|NIO|Swift|SPM|
 |-|-|-|-|
-|2.0|2.0|5.2+|`from: "2.0.0"`|
-|1.0|1.0|4.0+|`from: "1.0.0"`|
+|2.0|2.0|5.8+|`from: "2.0.0"`|
```
MahdiBM, line 27: isn't it incorrect to specify "5.8+" for "2.0.0"? Should just be 5.2+.

Package.swift
```diff
@@ -13,21 +13,38 @@ let package = Package(
...
     ]
 )
+
+var swiftSettings: [SwiftSetting] { [
+    .enableUpcomingFeature("ExistentialAny"),
+    .enableUpcomingFeature("ConciseMagicFile"),
+    .enableUpcomingFeature("ForwardTrailingClosures"),
+    .enableUpcomingFeature("DisableOutwardActorInference"),
+    .enableUpcomingFeature("StrictConcurrency"),
+    .enableExperimentalFeature("StrictConcurrency=complete"),
```
MahdiBM, line 49: Redundant? The experimental one should suffice?

MahdiBM (approved):
Looks good. Possibly add Dependabot too?

README.md
gwynne: This PR bumps the minimum requirement for PostgresKit to 5.8, so this is correct as is.

Package.swift
gwynne: The upcoming flag will replace the experimental one eventually, it's future-proofing.

README.md
MahdiBM: 2.0.0 still supports 5.8 though. You'll need to mention the correct number were requiring 5.8 started at. Like 2.14.

gwynne: 2.0 supported 5.1, so this was never correctly done from day one and I don't care enough to bother now any more than anyone other than you does 🙂

MahdiBM: sorry that first swift version was supposed to be 5.2 (although you said it's 5.1)

gwynne: There, happy now? :trollface:

MahdiBM (approved):

Package@swift-5.9.swift
0xTim, line 1: Do we actually have strict concurrency warnings without this manifest? I'm confused as to why we don't need this anymore, or was it because the flags weren't recognised in 5.7 but are in 5.8?

0xTim (approved):
Have you and @ptoffy been using the same for matter? 😆

Package@swift-5.9.swift
gwynne: Correct - the upcoming/experimental feature flag support was first available in SwiftPM as of 5.8; as of that version being the minimum, we don't need a separate manifest anymore unless we're using 5.9-only manifest features like macros.

PR #264, opened 2024-05-02
Override SQLDataType.blob to BYTEA for SQLKit users

gwynne:
Fluent already does this for users of FluentPostgresDriver, this just adds the same behavior to the equivalent data type at the SQLKit layer.


ptoffy:
Actually why is this crashing on the nightlies?

PR #268, opened 2024-07-05
Log query execution time.

maciejtrybilo:
I've added logging for query execution time.

This is quite useful when monitoring performance of endpoints that issue more than one database query.

### DIFF
--- a/Sources/PostgresKit/PostgresDatabase+SQL.swift
+++ b/Sources/PostgresKit/PostgresDatabase+SQL.swift
@@ -1,4 +1,5 @@
 import PostgresNIO
+import Dispatch
 import Logging
 import SQLKit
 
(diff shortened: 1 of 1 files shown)


maciejtrybilo:
Tracing is fairly complicated to set up. IMO this solution gives quite a lot for free.

I'm not sure how I could set something like this up on my side without having to instrument each call to Fluent or SQLKit?

MahdiBM:
Hmm I know @gwynne  is currently busy and might take a bit till she can answer, but i'd want to let her decide what we should do.

> gives quite a lot for free

The logging does have additional costs though. It's worth it to do if you need the info, but one might not.

> I'm not sure how I could set something like this up on my side without having to instrument each call to Fluent or SQLKit?

My Fluent knowledge is getting rusty ... but if you need such logging for Fluent queries or SQLKit queries, it might be better to have the log statements there, not in PostgresKit.

It also looks fine to me to have your own functions which instrument each call.

Again, I'll let Gwynne comment as she'd know better.

PR #275, opened 2025-10-23
Bump Swift minimum to 6.0

gwynne:
**These changes are now available in [2.15.0](https://github.com/vapor/postgres-kit/releases/tag/2.15.0)**

Also update CI and convert the tests to SwiftTesting.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,7 +59,7 @@ jobs:
     runs-on: ubuntu-latest
     container: ${{ matrix.swift-image }}
     services:
-      *postgres_host_a:
+      *postgres_host_a :
         image: ${{ matrix.postgres-image }}
         env:
           POSTGRES_USER: *postgres_user_a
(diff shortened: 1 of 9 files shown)


kylebrowning (approved):

.github/workflows/test.yml
```diff
@@ -42,29 +42,29 @@ jobs:
       fail-fast: false
       matrix:
         postgres-image:
-          - postgres:17
-          - postgres:15
-          - postgres:13
+          - postgres:18
+          - postgres:16
+          - postgres:14
```
MahdiBM, line 47: We should keep pg 13 I'd say, as long as Postgres themselves are supporting it.

gwynne: There were no protocol changes between 13 and 14, so I see no point in making the test matrix even larger; it doesn't actually tell you anything that the test for 14 doesn't.

MahdiBM: I wasnt thinking about a larger matrix. Mostly just replacing 14 with 13

PR #277, opened 2025-12-30
Ensure error messages thrown by data decoding provide useful information

gwynne:
**These changes are now available in [2.15.1](https://github.com/vapor/postgres-kit/releases/tag/2.15.1)**

Under some conditions, a failed attempt to decode data from an `SQLRow` (whether via SQLKit or through Fluent) would throw errors whose descriptions were missing information critical to understanding the problem. This fixes that.

Also adds some useful notes about using `PostgresKit` with `PostgresClient` to the README.


MahdiBM (approved):
LGTM

PR #279, opened 2026-07-18
Fix handling of array bindings containing optional values

gwynne:
**These changes are now available in [2.16.0](https://github.com/vapor/postgres-kit/releases/tag/2.16.0)**

- Now requires Swift 6.1
- We now use SQLKit's `SomeCodingKey` instead of redefining it.
- Encoding of `nil` binds to `NULL` is now on the fastest possible path instead of the slowest.
- Values that conform to `Postgres[Throwing]DynamicTypeEncodable` but not `Postgres[NonThrowing]Encodable` are now handled on the fast path.
- `nil` values in bound arrays having optional element types (which they should, as all arrays in Postgres have NULLable values) are now handled correctly instead of causing crashes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,15 @@
-<p align="center">
+<div align="center">
 <img src="https://design.vapor.codes/images/vapor-postgreskit.svg" height="96" alt="PostgresKit">
 <br>
-<br>
-<a href="https://docs.vapor.codes/4.0/"><img src="https://design.vapor.codes/images/readthedocs.svg" alt="Documentation"></a>
-<a href="https://discord.gg/vapor"><img src="https://design.vapor.codes/images/discordchat.svg" alt="Team Chat"></a>
-<a href="LICENSE"><img src="https://design.vapor.codes/images/mitlicense.svg" alt="MIT License"></a>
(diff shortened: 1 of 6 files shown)


ptoffy (approved):

MahdiBM (approved):

README.md
```diff
@@ -1,14 +1,15 @@
...
-<br>
-<a href="https://docs.vapor.codes/4.0/"><img src="https://design.vapor.codes/images/readthedocs.svg" alt="Documentation"></a>
-<a href="https://discord.gg/vapor"><img src="https://design.vapor.codes/images/discordchat.svg" alt="Team Chat"></a>
-<a href="LICENSE"><img src="https://design.vapor.codes/images/mitlicense.svg" alt="MIT License"></a>
-<a href="https://github.com/vapor/postgres-kit/actions/workflows/test.yml"><img src="https://img.shields.io/github/actions/workflow/status/vapor/postgres-kit/test.yml?event=push&style=plastic&logo=github&label=tests&logoColor=%23ccc" alt="Continuous Integration"></a>
-<a href="https://codecov.io/github/vapor/postgres-kit"><img src="https://img.shields.io/codecov/c/github/vapor/postgres-kit?style=plastic&logo=codecov&label=codecov" alt="Code Coverage"></a>
-<a href="https://swift.org"><img src="https://design.vapor.codes/images/swift60up.svg" alt="Swift 6.0+"></a>
-</p>
+
+[![Documentation](https://design.vapor.codes/images/readthedocs.svg)](https://docs.vapor.codes/4.0/)
```
0xTim, line 5: We can point to the API docs if that's better

0xTim (approved):

